PR #13, opened 2017-11-09
LWDEV-3794 Prevent running 2 instances of ME

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/AppStarter.kt
```diff
@@ -48,12 +58,13 @@ private fun teeLog(message: String) {
...
 
-internal class ShutdownHook : Thread() {
+internal class ShutdownHook(private val aliveStatusDatabaseAccessor: AliveStatusDatabaseAccessor) : Thread() {
     init {
         this.name = "ShutdownHook"
     }
 
     override fun run() {
         LOGGER.info("Stopping application")
+        aliveStatusDatabaseAccessor.unlock()
```
dmitrykrasovskih: reinit database

PR #29, opened 2017-12-15
LWDEV-4155 Pre-trade control

barsukovskii:
if resulting price differs from expected more than specified threshold - reject market order


src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -233,4 +243,16 @@ class MatchingEngine(private val LOGGER: Logger,
         val asset = if (order.isBuySide()) assetPair.quotingAssetId else assetPair.baseAssetId
         return balancesHolder.getAvailableBalance(order.clientId, asset)
     }
+
+    private fun checkPrice(order: NewOrder, price: Double, threshold: Double?): Boolean {
+        if (threshold == null) {
+            return true
+        }
+        val expectedPrice = order.expectedPrice() ?: return true
```
dmitrykrasovskih: add 0 check for expected price - we don't have control over incoming data

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -233,4 +243,16 @@ class MatchingEngine(private val LOGGER: Logger,
...
+        return if (order.isBuySide()) {
+            (price - expectedPrice) / expectedPrice < threshold
```
dmitrykrasovskih: check case if threshold = 0.0

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
```diff
@@ -93,13 +97,17 @@ class MarketOrderService(private val backOfficeDatabaseAccessor: BackOfficeDatab
...
 
-        val matchingResult = matchingEngine.match(order, orderBook)
+        val matchingResult = matchingEngine.match(order, orderBook, priceDifferenceThreshold = properties.priceDifferenceThreshold)
         val marketOrder = matchingResult.order as MarketOrder
         when (OrderStatus.valueOf(matchingResult.order.status)) {
             NoLiquidity -> {
                 rabbitSwapQueue.put(MarketOrderWithTrades(marketOrder))
                 writeResponse(messageWrapper, marketOrder, MessageStatus.NO_LIQUIDITY)
             }
+            PriceIsOutsideThreshold -> {
```
dmitrykrasovskih: add the same section to limit order services

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
barsukovskii: added

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
barsukovskii: added

dmitrykrasovskih:
outdated

PR #57, opened 2018-03-20
Lwdev 6315  read parameters from database

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/database/cache/ApplicationSettingsCache.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    private var settings: Settings = Settings()
+
+    override fun update() {
+        configDatabaseAccessor.loadConfigs()?.let { settings = it }
+    }
+
+    val trustedClients: Set<String>
```
dmitrykrasovskih: unify access to both propeties: rename method to isTrustedClient(client: String): Boolean

rocketmannot: fixed

PR #65, opened 2018-03-31
LVDEV-6337 dev add Spring to project

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/holders/BalancesHolder.kt
```diff
@@ -6,23 +6,31 @@ import com.lykke.matching.engine.database.WalletDatabaseAccessor
...
 
-    val wallets = walletDatabaseAccessor.loadWallets()
-    val initialClientsCount: Int = wallets.size
-    val initialBalancesCount: Int = wallets.values.sumBy { it.balances.size }
+
+    var wallets = walletDatabaseAccessor.loadWallets()
+    var initialClientsCount: Int = wallets.size
+    var initialBalancesCount: Int = wallets.values.sumBy { it.balances.size }
+
+    fun reload() {
```
dmitrykrasovskih: Is it possible to remove this method? It is used only from tests, and this is not good to add method only for tests.

src/main/kotlin/com/lykke/matching/engine/database/azure/AzureBackOfficeDatabaseAccessor.kt
```diff
@@ -3,21 +3,25 @@ package com.lykke.matching.engine.database.azure
...
 import com.microsoft.azure.storage.table.CloudTable
 import com.microsoft.azure.storage.table.TableOperation
 import com.microsoft.azure.storage.table.TableQuery
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
 import java.util.HashMap
 
-class AzureBackOfficeDatabaseAccessor(dictsConfig: String) : BackOfficeDatabaseAccessor {
+@Component
+class AzureBackOfficeDatabaseAccessor @Autowired constructor (config: Config) : BackOfficeDatabaseAccessor {
```
dmitrykrasovskih: Is it possible to continue using String and not whole config object?

src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/RabbitMqServiceDev.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.lykke.matching.engine.utils.config.RabbitConfig
+import org.apache.log4j.Logger
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Service
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.Executors
+
+@Service
+@Profile("dev")
+class RabbitMqServiceDev : RabbitMqService {
```
dmitrykrasovskih: Rename it to RabbitMqServiceToLogService

src/main/kotlin/com/lykke/matching/engine/services/HistoryTicksService.kt
```diff
@@ -1,19 +1,19 @@
...
-class HistoryTicksService(
-        private val marketStateCache: MarketStateCache,
-        private val genericLimitOrderService: GenericLimitOrderService,
-        private val frequency: Long) {
+class HistoryTicksService constructor (applicationContext: ApplicationContext,
+        private val genericLimitOrderService: GenericLimitOrderService) {
 
     private lateinit var historyTicksPersist: Timer
 
+    private var marketStateCache = applicationContext.getBean(MarketStateCache::class.java)
```
dmitrykrasovskih: why it is not @Autowired? or it will be changed after full migration?

src/main/kotlin/com/lykke/matching/engine/config/DevConfig.kt
```diff
@@ -0,0 +1,19 @@
+package com.lykke.matching.engine.config
+
+import com.lykke.matching.engine.utils.config.Config
+import com.lykke.utils.config.ConfigInitializer
+import org.springframework.beans.factory.FactoryBean
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+
+@Component
+@Profile("dev")
```
dmitrykrasovskih: Rename it to local config

src/main/kotlin/com/lykke/matching/engine/holders/BalancesHolder.kt
rocketmannot: previoiusly - in AbstractTests#initServices we were creating new instance of this class - and data was refreshed. Now Spring creates it once and data from database Accessor is not refetched when we need it, so i added this method.

src/main/kotlin/com/lykke/matching/engine/database/azure/AzureBackOfficeDatabaseAccessor.kt
rocketmannot: we can do something like 
```java
class AzureBackOfficeDatabaseAccessor @Autowired constructor (@Value("\${config.me.db.dictsConnString}") connectionString: String) : BackOfficeDatabaseAccessor {
```
But this is even worse - we just hardcoded in string name of the "config" bean.

src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/RabbitMqServiceDev.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/services/HistoryTicksService.kt
rocketmannot: some beans are hard to migrate at once - thay have deps that are not beans yet. This is temp solution- we will use annotation after migration.

src/main/kotlin/com/lykke/matching/engine/config/DevConfig.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/database/azure/AzureBackOfficeDatabaseAccessor.kt
rocketmannot: i thought about it once more. DatabaseAccessor should not know about Config. I will rewrite this. - only string will be supplied

rocketmannot:
merge not needed new branch was opened

PR #69, opened 2018-04-03
LWDEV-6594 History graps looks the same for 1h and 1d

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/database/azure/AzureHistoryTicksDatabaseAccessor.kt
```diff
@@ -45,14 +76,51 @@ class AzureHistoryTicksDatabaseAccessor(historyTicksString: String) : HistoryTic
...
-            val blob = historyBlobContainer.getBlockBlobReference(tick.name)
-            val byteArray = tick.convertToString().toByteArray()
-            blob.upload(ByteArrayInputStream(byteArray), byteArray.size.toLong())
-        } catch (e: Exception) {
-            LOGGER.error("Unable to save blob ${tick.name}", e)
-            METRICS_LOGGER.logError( "Unable to save blob ${tick.name}", e)
+    private fun blobToTick(blob: CloudBlob): TickBlobHolder? {
+        blob.downloadAttributes()
+        val names = blob.name.split("_")
+        val assetPair = names[1]
```
dmitrykrasovskih: asset pair will be incorrect if it is smth like BTCUSD_test, take asset pair name between first and last '_'

rocketmannot: sorry it is not clear for me, looks like this is version with fix from dev pull request -so in development  same version. in current implementation results will be flushed to file with correct name BA_USDBTC_1Y. My opinion is that we should ignore files with icorrect name - in prev version - we ignore them too.

PR #75, opened 2018-04-11
LWDEV-6704 mass LO cancel methods; LWDEV-6706 client multi limit order

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.log4j.Logger
+import java.util.Date
+
+class SingleLimitOrderProcessor(private val limitOrderService: GenericLimitOrderService,
+                                private val limitOrdersProcessorFactory: LimitOrdersProcessorFactory,
+                                private val assetsPairsHolder: AssetsPairsHolder,
+                                private val matchingEngine: MatchingEngine,
+                                private val LOGGER: Logger) {
+
+    fun processLimitOrder(messageWrapper: MessageWrapper?, order: NewLimitOrder, isCancelOrders: Boolean, now: Date, payBackReserved: Double?) {
```
rocketmannot: i suggest to split this method - add several overloads to this method.
1) one method that accapts messageWrapper
2) one without messageWrapper 
3) one for cancell orders only
In theory we should not have methods with boolean args and with optional params.

src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
```diff
@@ -0,0 +1,82 @@
...
+        val assetPair = assetsPairsHolder.getAssetPair(order.assetPairId)
+        val limitAsset = if (order.isBuySide()) assetPair.quotingAssetId else assetPair.baseAssetId
+        val orderBook = limitOrderService.getOrderBook(order.assetPairId).copy()
```
rocketmannot: may be getOrderBook should return copy always. I saw in many places where we called copy on returned order book to preserme immutability. Can we just return copu of order book all the time ?

src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
```diff
@@ -0,0 +1,82 @@
...
+        val clientsLimitOrdersWithTrades = mutableListOf<LimitOrderWithTrades>()
+        var buySideOrderBookChanged = false
+        var sellSideOrderBookChanged = false
+        var cancelVolume = 0.0
+        if (isCancelOrders) {
+            limitOrderService.getAllPreviousOrders(order.clientId, order.assetPairId, order.isBuySide()).forEach { orderToCancel ->
```
rocketmannot: Here i propose introducre two methods too.
getAllPreviousByingOrders and getAllPreviousBiddingOrders and drop last boolean argument

src/main/kotlin/com/lykke/matching/engine/order/cancel/AbstractLimitOrdersCanceller.kt
```diff
@@ -0,0 +1,188 @@
...
+            if (assetPair == null) {
+                ordersToRemove.getOrPut(order.assetPairId) { HashMap() }.getOrPut(order.isBuySide()) { LinkedList() }.add(order)
+            } else {
+                ordersToCancel.getOrPut(assetPair) { HashMap() }.getOrPut(order.isBuySide()) { LinkedList() }.add(order)
+            }
+        }
+        return preProcess(ordersToCancel, ordersToRemove)
+    }
+
+    fun preProcess(ordersToCancel: Map<AssetPair, Map<Boolean, Collection<NewLimitOrder>>>,
```
rocketmannot: may be here we need 4 Maps, ordersToCancell - for buy/bidding and orderToRemove for buy/bidding

src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
barsukovskii: Then in current case (and some other cases) we'll have to use if-operatior to call one of the methods.

src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
barsukovskii: In some cases we use orderbook only for reading. Then we don't need a copy.

src/main/kotlin/com/lykke/matching/engine/order/SingleLimitOrderProcessor.kt
barsukovskii: I switched messageWrapper to optional, default is null. 
But i didn't change cancellation. It is a part of method. During cancelleation we are calculating some data used in next process. And in case of splitting we should return this data and translate to 'process' method. I think we don't need it

PR #104, opened 2018-05-28
LWDEV-6712 Add message id field to incoming messages

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/services/MultiLimitOrderService.kt
```diff
@@ -85,14 +83,16 @@ class MultiLimitOrderService(private val limitOrderService: GenericLimitOrderSer
             parseMessage(messageWrapper)
         }
         if (!isOldTypeMessage) {
-            processMultiOrder(messageWrapper)
+            processClientOrders(messageWrapper)
```
barsukovskii: The method should be named 'processMultiOrder'. I guess it is merge mistake

src/main/kotlin/com/lykke/matching/engine/services/AbstractLimitOrdersCancelService.kt
```diff
@@ -40,10 +40,11 @@ abstract class AbstractLimitOrdersCancelService(protected val genericLimitOrderS
         val operationType = MessageType.valueOf(messageWrapper.type)?.name
                 ?: "Unknown message type ${messageWrapper.type}"
 
+        val operationId = messageWrapper.messageId ?: UUID.randomUUID().toString()
```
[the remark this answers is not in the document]
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/services/MultiLimitOrderService.kt
rocketmannot: fixed

PR #120, opened 2018-06-19
LWDEV-6992  Switch to BigDecimal

rocketmannot:


rocketmannot:
For reference see:
https://lykkex.atlassian.net/wiki/spaces/ME/pages/500989968/Strategy+of+BigDecimal+usage+in+matching+engine?src=jira

PR #136, opened 2018-07-12
LWDEV-7960 Check total balance for trusted clients

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -0,0 +1,108 @@
+package com.lykke.matching.engine.services.utils
+
+import com.lykke.matching.engine.daos.LimitOrder
+import com.lykke.matching.engine.utils.NumberUtils
+import org.apache.log4j.Logger
+import java.math.BigDecimal
+
+class MultiOrderFilter(private val isTrustedClient: Boolean,
```
rocketmannot: my general suggestion is to 
1) make this class stateless and in future give priority to stateless classes.
2) sort volume-prices - and process in filter already sorted list - in the future we will have cotnext  that will contain already sorted volume-prices. Context will be initialized in separate - non business thread - so business tread will not suffer from bad performance- in such a way we will make implementation of this logic more clear

rocketmannot: Decided to leave it as is -  refacoring is planned after implementation of context.

PR #150, opened 2018-07-31
LWDEV-8497 Do not use redis pool

barsukovskii:


rocketmannot:
now we are using persistence manager in preprocessor thread - to persist processed messages. So we are using same jedis instance in several threads - previosly jedis pool was used in persistence manager - we need more investigation on the topic now.

PR #151, opened 2018-07-31
LWDEV-8443 Prevent trades with 0 volume

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -125,15 +129,29 @@ class MatchingEngine(private val LOGGER: Logger,
...
                         limitOrder,
                         if (isBuy) marketRoundedVolume else oppositeRoundedVolume,
                         limitReservedBalances)) {
-                    LOGGER.info("Added order (id: ${limitOrder.externalId}, client: ${limitOrder.clientId}, asset: ${limitOrder.assetPairId}) to cancelled limit orders")
+                    LOGGER.info("Added order ($limitOrderInfo) to cancelled limit orders")
                     cancelledLimitOrders.add(limitOrderCopyWrapper)
                     continue
                 }
 
+                if (BigDecimal.ZERO.compareTo(if (isBuy) marketRoundedVolume else oppositeRoundedVolume) == 0) {
```
rocketmannot: use NumberUtils.equalsIgnoreScale

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -269,8 +287,13 @@ class MatchingEngine(private val LOGGER: Logger,
         }
 
         if (order.takePrice() == null && remainingVolume > BigDecimal.ZERO) {
```
rocketmannot: by checking order.takerPrice - do we test if this is market order ? suggest to use instanceOf - it is not so slow on target JVM 1.8. Also suggest to remove parameter price in MarketOrder and always pass null in Order underlying constructor.

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
barsukovskii: We can't remove field 'price' - it's used in outgoing messages.
We can only remove parameter from constructor because market order price is lateinit field.
But it is another task

PR #154, opened 2018-08-03
LWDEV-8647 Pre-trade control: max order volume and value

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -291,6 +291,31 @@ class MatchingEngine(private val LOGGER: Logger,
...
+        val executionPrice = NumberUtils.setScale(
+                if (order.isStraight())
+                    NumberUtils.divideWithMaxScale(totalLimitPrice, order.getAbsVolume())
+                else
+                    NumberUtils.divideWithMaxScale(order.getAbsVolume(), totalVolume),
+                assetPair.accuracy,
+                order.isOrigBuySide()
+        )
+
+        if (order.takePrice() == null) {
```
rocketmannot: suggest to move this code out of match method to separate method - it is already too large and confused. For example create two new methods - getExecutionPrice() and check maxVolume()

src/main/kotlin/com/lykke/matching/engine/services/validators/impl/MarketOrderValidatorImpl.kt
```diff
@@ -55,10 +55,18 @@ class MarketOrderValidatorImpl
...
             LOGGER.info("Too small volume for $order")
             throw OrderValidationException(OrderStatus.TooSmallVolume)
         }
+        if (order.isStraight() && assetPair.maxVolume != null && order.getAbsVolume() > assetPair.maxVolume) {
+            LOGGER.info("Too large volume for $order")
+            throw OrderValidationException(OrderStatus.InvalidVolume)
+        }
+        if (!order.isStraight() && assetPair.maxValue != null && order.getAbsVolume() > assetPair.maxValue) {
+            LOGGER.info("Too large value for $order")
+            throw OrderValidationException(OrderStatus.InvalidVolume)
```
rocketmannot: Here as i suggested may be we need separate status - InvalidValue - this that we return InvalidVolume when we have invalid value - this can be confusing for clients.

src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
barsukovskii: Fixed

src/main/kotlin/com/lykke/matching/engine/messages/MessageStatus.kt
```diff
@@ -23,5 +23,6 @@ enum class MessageStatus(val type: Int){
     NO_LIQUIDITY(411),
     NOT_ENOUGH_FUNDS(412),
     INVALID_VOLUME_ACCURACY(431),
-    INVALID_PRICE_ACCURACY(432)
+    INVALID_PRICE_ACCURACY(432),
+    INVALID_VOLUME(434)
```
dmitrykrasovskih: Let's add new status, INVALID_ORDER_VALUE

src/main/kotlin/com/lykke/matching/engine/services/validators/impl/MarketOrderValidatorImpl.kt
barsukovskii: Added new status

src/main/kotlin/com/lykke/matching/engine/messages/MessageStatus.kt
barsukovskii: Done

PR #166, opened 2018-08-23
LWDEV-8998 Use Spring mvc in httpOrderBook endpoint

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/RabbitMqPublisher.kt
```diff
@@ -0,0 +1,60 @@
...
+                        appVersion: String,
+                        exchangeType: BuiltinExchangeType,
+                        private val jsonMessageTransformer: LogMessageTransformer,
+                        /** null if do not need to log */
+                        messageDatabaseLogger: DatabaseLogger? = null) : AbstractRabbitMqPublisher<Event<*>>(uri, exchangeName,
+        queue, appName, appVersion, exchangeType, LOGGER,
+        MESSAGES_LOGGER, METRICS_LOGGER, STATS_LOGGER, messageDatabaseLogger) {
+
+    companion object {
+        private val LOGGER = ThrottlingLogger.getLogger(RabbitMqOldFormatPublisher::class.java.name)
```
barsukovskii: wrong logger names

rocketmannot: fixed

PR #172, opened 2018-08-31
LWDEV-9067 Add swagger methods to change trusted clients and disabled assets

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/database/azure/AzureSettingsDatabaseAccessor.kt
```diff
@@ -0,0 +1,154 @@
...
+            settingsTable.execute(TableOperation.delete(it))
+        }
+    }
+
+    override fun deleteSettingsGroup(settingGroupName: String) {
+        val settings = getAzureSettingsForGroup(settingGroupName)
+        batchDelete(settingsTable, settings)
+    }
+
+    override fun createOrUpdateSetting(settingGroupName: String, setting: Setting) {
```
barsukovskii: Return flag (or exception in case of persistence error) to know operation result. Write changes to cache when result is OK. The same for other editing methods (and perhaps also for SettingsHistoryDatabaseAccessor)

src/main/kotlin/com/lykke/matching/engine/web/dto/SettingDto.kt
```diff
@@ -0,0 +1,15 @@
+package com.lykke.matching.engine.web.dto
+
+import io.swagger.annotations.ApiModelProperty
+import java.util.*
+import javax.validation.constraints.NotNull
+
+data class SettingDto(@get:NotNull(message = "Name should not be null") val name: String,
+                      @get:NotNull(message = "Value should not be null") val value: String,
+                      @get:NotNull(message = "Enabled flag should not be null") val enabled: Boolean?,
```
barsukovskii: The filed is marked as NotNull, but 'Boolean?'

rocketmannot: i made it nullable specially for this case, if user pass no value for flag - we should return error

src/main/java/com/lykke/matching/engine/daos/azure/settings/AzureAppSetting.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.microsoft.azure.storage.table.TableServiceEntity;
+
+public class AzureAppSetting extends TableServiceEntity {
+
+    private String value;
+
+    private String comment;
+
+    private String user;
```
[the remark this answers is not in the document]
rocketmannot: moved to history entity only

PR #173, opened 2018-09-04
LWDEV-9189 Write statuses of filtered out orders to socket for truste…

barsukovskii:
…d client multi-limit orders

### DIFF
--- a/src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
+++ b/src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
@@ -1,14 +1,17 @@
 package com.lykke.matching.engine.services.utils
 
 import com.lykke.matching.engine.daos.LimitOrder
+import com.lykke.matching.engine.order.OrderStatus
 import com.lykke.matching.engine.utils.NumberUtils
 import org.apache.log4j.Logger
 import java.math.BigDecimal
+import java.util.Date
 
 class MultiOrderFilter(private val isTrustedClient: Boolean,
                        private val baseAssetAvailableBalance: BigDecimal,
                        private val quotingAssetAvailableBalance: BigDecimal,
                        private val quotingAssetAccuracy: Int,
+                       private val date: Date,
                        initialCapacity: Int,
                        private val LOGGER: Logger) {
 
@@ -20,18 +23,18 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
     private var usedQuotingAssetVolume = BigDecimal.ZERO
     private val orders = ArrayList<LimitOrder>(initialCapacity)
 
-    fun checkAndAdd(order: LimitOrder): Boolean {
+    fun checkAndAdd(order: LimitOrder) {
+        orders.add(order)
         if (!isTrustedClient) {
-            return orders.add(order)
+            return
         }
-        var skip = false
         if (order.isBuySide()) {
             if (!notSortedBuySide) {
                 if (prevBidPrice == null || order.price < prevBidPrice) {
                     prevBidPrice = order.price
                     val volume = NumberUtils.setScaleRoundUp(order.volume * order.price, quotingAssetAccuracy)
                     if (usedQuotingAssetVolume + volume > quotingAssetAvailableBalance) {
-                        skip = true
+                        order.updateStatus(OrderStatus.NotEnoughFunds, date)
                         LOGGER.info("[${order.assetPairId}] Unable to add order ${order.volume} @ ${order.price} (${order.externalId}) due to low balance (available: $quotingAssetAvailableBalance, used: $usedQuotingAssetVolume)")
                     } else {
                         usedQuotingAssetVolume += volume
@@ -47,7 +50,7 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
                     prevAskPrice = order.price
                     val volume = order.getAbsVolume()
                     if (usedBaseAssetVolume + volume > baseAssetAvailableBalance) {
-                        skip = true
+                        order.updateStatus(OrderStatus.NotEnoughFunds, date)
                         LOGGER.info("[${order.assetPairId}] Unable to add order ${order.volume} @ ${order.price} (${order.externalId}) due to low balance (available: $baseAssetAvailableBalance, used: $usedBaseAssetVolume)")
                     } else {
                         usedBaseAssetVolume += volume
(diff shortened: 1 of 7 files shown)


src/test/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilterTest.kt
```diff
@@ -19,72 +21,91 @@ class MultiOrderFilterTest {
...
-            sortedOrders[3],
-            sortedOrders[8],
-            sortedOrders[7],
-            sortedOrders[6],
-            sortedOrders[5])
+    private fun notSortedOrders(): List<LimitOrder> {
+        val sortedOrders = sortedOrders()
+        return listOf(sortedOrders[0].copy(),
+                sortedOrders[1].copy(),
+                sortedOrders[2].copy(),
```
rocketmannot: can we shuffle randomly this items automatically ?

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -77,31 +78,26 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
         }
         buyOrders.sortWith(Comparator { order1, order2 -> -order1.price.compareTo(order2.price) })
```
rocketmannot: may be we can use sortedBy(order.price) and sortByDescending(order.price) ?

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -58,16 +61,14 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
                 }
             }
         }
-
-        return if (!skip) orders.add(order) else false
     }
 
-    fun filterOutIfNotSorted(): List<LimitOrder> {
+    fun checkIfNotSorted() {
```
rocketmannot: suggest to remove this method and perform computation on getResult- first time and than cache computed value - thus we will not miss to call checkIfNotSorted in future uses of filter

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -20,18 +23,18 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
...
-            return orders.add(order)
+            return
         }
-        var skip = false
         if (order.isBuySide()) {
             if (!notSortedBuySide) {
                 if (prevBidPrice == null || order.price < prevBidPrice) {
                     prevBidPrice = order.price
                     val volume = NumberUtils.setScaleRoundUp(order.volume * order.price, quotingAssetAccuracy)
                     if (usedQuotingAssetVolume + volume > quotingAssetAvailableBalance) {
```
rocketmannot: we have here 4 levels of if blocks - sugges to decompose these blocks on names functions - thus it will be easier to read code, also as i can see buy and sell cases have almost similar logic - can we deduplicate it ?

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -20,18 +23,18 @@ class MultiOrderFilter(private val isTrustedClient: Boolean,
...
         if (!isTrustedClient) {
-            return orders.add(order)
+            return
         }
-        var skip = false
         if (order.isBuySide()) {
             if (!notSortedBuySide) {
                 if (prevBidPrice == null || order.price < prevBidPrice) {
                     prevBidPrice = order.price
                     val volume = NumberUtils.setScaleRoundUp(order.volume * order.price, quotingAssetAccuracy)
```
rocketmannot: suggest to rename to value and name ( volume * price) as  "value" across project in future

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
```diff
@@ -1,14 +1,17 @@
...
 import com.lykke.matching.engine.utils.NumberUtils
 import org.apache.log4j.Logger
 import java.math.BigDecimal
+import java.util.Date
 
 class MultiOrderFilter(private val isTrustedClient: Boolean,
                        private val baseAssetAvailableBalance: BigDecimal,
                        private val quotingAssetAvailableBalance: BigDecimal,
                        private val quotingAssetAccuracy: Int,
+                       private val date: Date,
```
rocketmannot: why we need to pass now instead of generation it inside filter ?  please set more meaning name to the parameter "date" - is too generic

src/test/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilterTest.kt
barsukovskii: This could not work in the previous implementation of this test, but works now. Changed.

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
barsukovskii: Yes. Fixed

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
barsukovskii: Fixed. Added an exception throwing in case 'checkAndAdd' is invoked after 'getResult' invocation

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
barsukovskii: Fixed

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
barsukovskii: renamed to 'quotingAssetVolume'

src/main/kotlin/com/lykke/matching/engine/services/utils/MultiOrderFilter.kt
barsukovskii: Status date must be equal to execution time

PR #181, opened 2018-09-17
LWDEV-9021 Add swagger method to get all open orders for client

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/web/controllers/ClientOrdersController.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    @Autowired
+    private lateinit var genericStopLimitOrderService: GenericStopLimitOrderService
+
+    @GetMapping("client/{clientId}/orders", produces = [MediaType.APPLICATION_JSON_VALUE])
+    @ApiOperation("Endpoint to access orders from current order book for given client")
+    fun getOrders(@PathVariable("clientId") clientId: String,
+                  @RequestParam(required = false) assetPair: String?,
+                  @RequestParam(required = false) isBuy: Boolean?): ClientOrdersDto {
+        val limitOrders = genericLimitOrderService.searchOrders(clientId, assetPair, isBuy)
```
barsukovskii: this call can produce ConcurrentModificationException: the method loops mutable list which can be changed in another thread

rocketmannot: This method will be used only for debug reasons. it is known trade off to omit synchronization - exception will be thrown in this calling thread - so only REST api thread will suffer - this is ok

PR #186, opened 2018-09-20
LWDEV-8998 Use Spring mvc in httpOrderBook endpoint

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/RabbitMqPublisher.kt
```diff
@@ -0,0 +1,64 @@
...
+
+        // MINIMAL_PERSISTENT_BASIC + headers
+        return AMQP.BasicProperties.Builder()
+                .deliveryMode(2)
+                .headers(headers)
+                .build()
+    }
+
+    private fun getLogMessage(item: Event<*>): String {
+        return gson.toJson(item)
```
barsukovskii: Serialize to json only if it is needed to log (messageDatabaseLogger is not null)

src/main/kotlin/com/lykke/matching/engine/web/dto/StopOrderBook.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.util.*
+
+data class StopOrderBook(val assetPair: String,
+                         @get:JsonProperty("isBuy")
+                         @JsonProperty("isBuy")
+                         val isBuy: Boolean,
+
+                         @get:JsonProperty("isLower")
```
barsukovskii: Is it possible to configure boolean fields names policy during gson initialization ?

rocketmannot: this is greate hack - to display isBuy in swagger examples - as springfox harcoded jackson  json serializer in their code - so this is annotation for jackson

src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/RabbitMqPublisher.kt
rocketmannot: fixed

PR #189, opened 2018-09-24
LWDEV-9400 add validation for stop orders

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/order/StopLimitOrderProcessor.kt
```diff
@@ -200,6 +200,7 @@ class StopLimitOrderProcessor(private val limitOrderService: GenericLimitOrderSe
         validator.validateAssets(assetPair)
         validator.validateLimitPrices(order)
         validator.validateVolume(order, assetPair)
+        validator.validateMaxValue(order, assetPair)
```
barsukovskii: Let's rename this method to validateStopOrderMaxValue, because method 'validateVolume' already has max value validation for limit order, and it may confuse

rocketmannot: fixed

PR #204, opened 2018-10-11
LWDEV-9418 Multi limit order asset pair validation

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/services/MultiLimitOrderService.kt
```diff
@@ -449,6 +448,12 @@ class MultiLimitOrderService(private val limitOrderService: GenericLimitOrderSer
                 (if (message.hasCancelAllPreviousLimitOrders()) "cancelPrevious: ${message.cancelAllPreviousLimitOrders}, " else "") +
                 (if (message.hasCancelMode()) "cancelMode: ${message.cancelMode}" else ""))
 
+        val assetPair = assetsPairsHolder.getAssetPairAllowNulls(message.assetPairId)
```
rocketmannot: as far as i can see - this is fix for old orders only ? for new we still can get exception. may be we need to add validation to MultilimitOrderValidator too?

barsukovskii: No, this fix is only for new orders.

PR #206, opened 2018-10-11
LWDEV-9477 Remove OLD_MARKET_ORDER obsolete message processing

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
```diff
@@ -288,8 +288,6 @@ class MessageProcessor(config: Config, messageRouter: MessageRouter, application
         result[MessageType.LIMIT_ORDER] = singleLimitOrderService
         result[MessageType.OLD_LIMIT_ORDER] = singleLimitOrderService
         result[MessageType.MARKET_ORDER] = marketOrderService
-        result[MessageType.NEW_MARKET_ORDER] = marketOrderService
```
rocketmannot: why new_market_order is removed ?

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
```diff
@@ -312,59 +296,26 @@ class MarketOrderService @Autowired constructor(
...
-                    .setStatus(status.type)
-
-            if (order.price != null) {
-                marketOrderResponse.price = order.price!!.toDouble()
-            } else if (reason != null) {
-                marketOrderResponse.statusReason = reason
-            }
-
-            messageWrapper.writeMarketOrderResponse(marketOrderResponse)
+        val marketOrderResponse = ProtocolMessages.MarketOrderResponse.newBuilder()
```
rocketmannot: if i understand correctly previously we were writing these reponses:
OLD_MARKET_ORDER => Response
MARKET_ORDER => NewResponse
NEW_MARKET_ORDER => MarketOrderResponse

Now
 MARKET_ORDER => MarketOrderResponse instead of NewResponse
is this correct?

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
```diff
@@ -312,59 +296,26 @@ class MarketOrderService @Autowired constructor(
...
-        if (messageWrapper.type == MessageType.OLD_MARKET_ORDER.type) {
-            messageWrapper.writeResponse(ProtocolMessages.Response.newBuilder())
-        } else if (messageWrapper.type == MessageType.MARKET_ORDER.type) {
-            messageWrapper.writeNewResponse(ProtocolMessages.NewResponse.newBuilder()
-                    .setStatus(status.type))
-        } else {
-            messageWrapper.writeMarketOrderResponse(ProtocolMessages.MarketOrderResponse.newBuilder()
-                    .setStatus(status.type))
-        }
+        messageWrapper.writeMarketOrderResponse(ProtocolMessages.MarketOrderResponse.newBuilder()
```
rocketmannot: The same as above should we write here NewResponse?

src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
barsukovskii: Code of MessageType.MARKET_ORDER was switched to 53, so now we support only type=53 and send "MARKET_ORDER" in rmq event as it was before

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
barsukovskii: Yes, the same: code of MessageType.MARKET_ORDER was switched to 53

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
barsukovskii: Yes, it's OK

PR #209, opened 2018-10-11
LWDEV-9665 Turn on maintenance mode if at least one of input queues size reaches threshold	

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/utils/config/MatchingEngineConfig.kt
```diff
@@ -20,6 +20,7 @@ data class MatchingEngineConfig(
         val candleSaverInterval: Long,
         val hoursCandleSaverInterval: Long,
         val queueSizeLoggerInterval: Long,
+        val queueSizeHealthCheckInterval: Long,
```
dmitrykrasovskih: Move all queue related params to separate sub-config

rocketmannot: fixed

PR #211, opened 2018-10-17
LWDEV-9731 Turn on maintenance mode when rabbit mq connections are lost

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/services/MessageSender.kt
```diff
@@ -7,7 +7,7 @@ import java.util.concurrent.BlockingQueue
 
 @Component
 class MessageSender(private val clientsEventsQueue: BlockingQueue<Event<*>>,
-                    private val trustedClientsEventsQueue: BlockingQueue<ExecutionEvent>) {
+                    private val trustedClientsEventsQueue: BlockingQueue<Event<*>>) {
```
barsukovskii: Why is this changed? only ExecutionEvent can be trusted client event

src/main/kotlin/com/lykke/matching/engine/outgoing/rabbit/impl/publishers/AbstractRabbitMqPublisher.kt
```diff
@@ -10,19 +12,23 @@ import com.rabbitmq.client.Channel
...
-                                               private val exchangeType: BuiltinExchangeType,
-                                               private val LOGGER: ThrottlingLogger,
-                                               private val MESSAGES_LOGGER: Logger,
-                                               private val METRICS_LOGGER: MetricsLogger,
-                                               private val STATS_LOGGER: Logger,
-                                               /** null if do not need to log */
+abstract class AbstractRabbitMqPublisher<T>(private val uri: String,
+                                            private val exchangeName: String,
+                                            private val queueName: String,
+                                            val queue: BlockingQueue<out T>,
```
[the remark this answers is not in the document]
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/services/MessageSender.kt
rocketmannot: fixed

PR #212, opened 2018-10-17
LWDEV-9861 Multi stop limit orders; stop orders transaction

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionBalancesHolder.kt
```diff
@@ -1,13 +1,12 @@
...
 import com.lykke.matching.engine.daos.wallet.Wallet
 import com.lykke.matching.engine.database.common.entity.BalancesData
-import com.lykke.matching.engine.database.common.entity.PersistenceData
 import com.lykke.matching.engine.holders.BalancesHolder
 import java.math.BigDecimal
 
-class BalancesUpdater(private val balancesHolder: BalancesHolder) {
+class CurrentTransactionBalancesHolder(private val balancesHolder: BalancesHolder) {
 
     private val changedBalances = mutableMapOf<String, MutableMap<String, AssetBalance>>()
```
rocketmannot: suggest to use name format for maps keyToValue

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
```diff
@@ -23,35 +25,33 @@ import java.util.Date
...
 
     companion object {
         private val LOGGER = Logger.getLogger(WalletOperationsProcessor::class.java.name)
         private val METRICS_LOGGER = MetricsLogger.getLogger()
     }
 
-    private val balancesUpdater = balancesHolder.createUpdater()
-    private val changedAssetBalances = HashMap<String, ChangedAssetBalance>()
     private val clientIds = HashSet<String>()
     private val updates = HashMap<String, ClientBalanceUpdate>()
```
rocketmannot: please name it rabbitClientBalanceUpdates

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContext.kt
```diff
@@ -0,0 +1,89 @@
...
+                       val date: Date,
+                       val logger: Logger) {
+
+    companion object {
+        private val DEBUGGER = Logger.getLogger(ExecutionContext::class.java.name)
+    }
+
+    var tradeIndex: Long = 0
+
+    private val clientLimitOrdersWithTradesByInternalId = LinkedHashMap<String, LimitOrderWithTrades>()
```
rocketmannot: please use keyToValuePattern to name maps.

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionBalancesHolder.kt
barsukovskii: Fixed. Discussed and decided to use format valuesByKey.

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
barsukovskii: Fixed. Renamed to rabbitClientBalanceUpdatesByClientIdAndAssetId according to map name format

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContext.kt
barsukovskii: Decided to use format valuesByKey

src/main/kotlin/com/lykke/matching/engine/services/GenericStopLimitOrderService.kt
```diff
@@ -90,6 +67,16 @@ class GenericStopLimitOrderService(private val stopOrdersDatabaseAccessorsHolder
         }
     }
 
+    override fun removeOrderFromMapsWithStatus(orders: Collection<LimitOrder>, status: OrderStatus?, date: Date?) {
```
rocketmannot: confusing method - sounds like we are removing orders with given status - but we are updating status of removed order

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContextFactory.kt
```diff
@@ -0,0 +1,59 @@
...
+                assetsById,
+                preProcessorValidationResultsByOrderId,
+                walletOperationsProcessor,
+                orderBooksHolder,
+                stopOrderBooksHolder,
+                date,
+                logger)
+    }
+
+    private fun getAssetsById(assetPairsById: Map<String, AssetPair>): Map<String, Asset> {
```
rocketmannot: suggest to rename to getAssetsByIdMap

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionOrderBooksHolder.kt
```diff
@@ -0,0 +1,93 @@
...
+                orderBookPersistenceDataList.add(OrderBookPersistenceData(assetPairId, false, updatedOrders.updatedOrderBook))
+                updatedOrders.updatedOrder?.let { ordersToSave.add(it) }
+            }
+        }
+
+        ordersToSave.addAll(newOrdersByExternalId.values)
+        return OrderBooksPersistenceData(orderBookPersistenceDataList, ordersToSave, ordersToRemove)
+    }
+
+    private fun getUpdatedOrderBookAndOrder(orderBook: PriorityBlockingQueue<LimitOrder>): UpdatedOrderBookAndOrder {
```
rocketmannot: This method is confusing, looks like it name implies that this is getter method - which should not mutate data, however inside of method we are mutating orderBook. Also it is not clear why are we returning orderBook with changed order and using only changed order in client code

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionBalancesHolder.kt
```diff
@@ -1,16 +1,15 @@
...
-import com.lykke.matching.engine.database.common.entity.PersistenceData
 import com.lykke.matching.engine.holders.BalancesHolder
 import java.math.BigDecimal
 
-class BalancesUpdater(private val balancesHolder: BalancesHolder) {
+class CurrentTransactionBalancesHolder(private val balancesHolder: BalancesHolder) {
 
-    private val changedBalances = mutableMapOf<String, MutableMap<String, AssetBalance>>()
-    private val changedWallets = mutableMapOf<String, Wallet>()
+    private val changedBalancesByClientIdAndAssetId = mutableMapOf<String, MutableMap<String, AssetBalance>>()
```
rocketmannot: why do we need this map. We already have assetBalancesData in `changedWalletsByClientId`

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
```diff
@@ -0,0 +1,80 @@
...
+        addRemovedOrders(orders, cancelledOrders)
+    }
+
+    fun addReplacedOrders(orders: Collection<LimitOrder>) {
+        addRemovedOrders(orders, replacedOrders)
+    }
+
+    fun apply(date: Date) {
+        genericLimitOrderService.removeOrderFromMapsWithStatus(completedOrders)
+        genericLimitOrderService.removeOrderFromMapsWithStatus(cancelledOrders, OrderStatus.Cancelled, date)
```
rocketmannot: Confusing name for method `removeOrderFromMapsWithStatus` - for first sign we are removing orders with given status, but in fact we are removing order and setting status. Also `removeOrderFromMapsWithStatus` has to responsibilities remove order and set status, please separate these two responsibilities

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
```diff
@@ -0,0 +1,80 @@
...
+        assetOrderBookCopiesByAssetPairId[orderBook.assetPairId] = orderBook
+    }
+
+    fun addOrder(order: LimitOrder) {
+        getChangedOrderBookCopy(order.assetPairId).addOrder(order)
+        newOrdersByExternalId[order.externalId] = order
+        addChangedSide(order)
+    }
+
+    fun addCompletedOrders(orders: Collection<LimitOrder>) {
```
rocketmannot: this method is not atomic as well

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Suppress("unchecked_cast")
+    fun getChangedOrderBookCopy(assetPairId: String): AssetOrderBook {
+        return assetOrderBookCopiesByAssetPairId.getOrPut(assetPairId) {
+            genericLimitOrderService.getOrderBook(assetPairId).copy() as AssetOrderBook
+        }
+    }
+
+    fun setOrderBook(orderBook: AssetOrderBook) {
```
rocketmannot: This method is error prone. As far as i understand it is assumed that after calling it user of this api will also call `addCancelledOrders` or `addCompletedOrders` to preserve inner invariants of this class, in fact new user can forget to do that. The desired api for this class should be atomic- after call of any method of the class data should stay valid.

src/main/kotlin/com/lykke/matching/engine/order/process/StopOrderBookProcessor.kt
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+        return null
+    }
+
+    private fun getStopOrderToExecuteByAssetPair(assetPairId: String, executionContext: ExecutionContext): LimitOrder? {
+        val orderBook = executionContext.orderBooksHolder.getChangedCopyOrOriginalOrderBook(assetPairId)
+        val bestBidPrice = orderBook.getBidPrice()
+        val bestAskPrice = orderBook.getAskPrice()
+        val order = executionContext.stopOrderBooksHolder.pollStopOrderToExecute(assetPairId,
```
rocketmannot: suggest to rename to 'get', as poll implies we are waiting some timeout for stop order

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
```diff
@@ -0,0 +1,322 @@
...
+            matchingResultHandlingHelper.preProcessCancelledOppositeOrders(orderContext)
+        }
+
+        if (matchingResult.uncompletedLimitOrderCopy != null) {
+            matchingResultHandlingHelper.preProcessUncompletedOppositeOrder(orderContext)
+        }
+
+        orderContext.ownWalletOperations = matchingResult.ownCashMovements
+        val orderCopy = matchingResult.order as LimitOrder
+        if (orderCopy.status == OrderStatus.Processing.name || orderCopy.status == OrderStatus.InOrderBook.name) {
```
rocketmannot: predicate is duplicated - suggest to extract to separate method

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
```diff
@@ -0,0 +1,322 @@
...
+        val order = orderContext.order
+        val orderBook = executionContext.orderBooksHolder.getChangedCopyOrOriginalOrderBook(order.assetPairId)
+        val matchingResult = matchingEngine.match(order,
+                orderBook.getOrderBook(!order.isBuySide()),
+                executionContext.messageId,
+                orderContext.availableLimitAssetBalance!!,
+                applicationSettingsCache.limitOrderPriceDeviationThreshold(order.assetPairId),
+                executionContext = executionContext)
+        orderContext.matchingResult = matchingResult
+        val orderCopy = matchingResult.order as LimitOrder
```
rocketmannot: suggest to rename `matchingResult.order` to `matchingResult.orderCopy`

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
```diff
@@ -0,0 +1,322 @@
...
+            return OrderValidationResult(false, false, e.message, e.orderStatus)
+        }
+        orderContext.availableLimitAssetBalance = availableLimitAssetBalance
+        return OrderValidationResult(true)
+    }
+
+    private fun processInvalidOrder(orderContext: LimitOrderExecutionContext): ProcessedOrder {
+        val order = orderContext.order
+        val validationResult = orderContext.validationResult!!
+        orderContext.executionContext.info("${getOrderInfo(order)} is rejected: ${validationResult.message}")
```
rocketmannot: should we user here error  level, In general this is bad idea, but as we have agreed to use error level for all rejections

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
```diff
@@ -0,0 +1,322 @@
...
+        try {
+            limitOrderBusinessValidator.performValidation(applicationSettingsCache.isTrustedClient(order.clientId),
+                    order,
+                    availableLimitAssetBalance,
+                    orderContext.limitVolume!!,
+                    orderContext.executionContext.orderBooksHolder.getChangedCopyOrOriginalOrderBook(order.assetPairId))
+        } catch (e: OrderValidationException) {
+            return OrderValidationResult(false, false, e.message, e.orderStatus)
+        }
+        orderContext.availableLimitAssetBalance = availableLimitAssetBalance
```
rocketmannot: validation method is supposed to not change state, please extract this mutation out of validation method

src/main/kotlin/com/lykke/matching/engine/order/cancel/GenericLimitOrdersCanceller.kt
```diff
@@ -84,73 +73,53 @@ class GenericLimitOrdersCanceller(dictionariesDatabaseAccessor: DictionariesData
...
-                    limitOrdersWithTrades))
-        }
+        stopOrderBookProcessor.checkAndExecuteStopLimitOrders(executionContext)
 
-        limitOrdersCanceller.checkAndProcessStopOrders(messageId)
+        return executionConfirmationService.persistAndSendEvents(messageWrapper, executionContext)
+    }
 
-        return true
+    private fun getAssetPairsById(assetPairIds: Collection<String>): Map<String, AssetPair> {
```
rocketmannot: suggest to rename to `getAssetPairsByIdMap`

src/main/kotlin/com/lykke/matching/engine/order/cancel/AbstractLimitOrdersCanceller.kt
```diff
@@ -151,20 +133,6 @@ abstract class AbstractLimitOrdersCanceller<TAssetOrderBook : AbstractAssetOrder
                                            trustedClientsOrdersWithTrades: List<LimitOrderWithTrades>,
                                            assetOrderBooks: Map<String, TAssetOrderBook>): TCancelResult
 
-    open fun apply(messageId: String,
```
rocketmannot: now we have AbstractLimitOrdersCanceller is not applying data to cache, so why we need preProcess/process steps here, can we just have one process step ?

src/main/kotlin/com/lykke/matching/engine/order/cancel/AbstractLimitOrdersCanceller.kt
```diff
@@ -18,6 +15,7 @@ import com.lykke.matching.engine.services.utils.AbstractAssetOrderBook
 import java.math.BigDecimal
 import java.util.*
 import java.util.stream.Collectors
+import kotlin.collections.ArrayList
```
rocketmannot: redundant import

src/main/kotlin/com/lykke/matching/engine/order/ExecutionEventSender.kt
```diff
@@ -0,0 +1,96 @@
...
+                    executionContext.requestId,
+                    executionContext.date,
+                    executionContext.messageType,
+                    executionContext.walletOperationsProcessor.getClientBalanceUpdates(),
+                    clientsLimitOrdersWithTrades,
+                    executionContext.marketOrderWithTrades))
+        }
+    }
+
+    private fun isTrustedClientEvent(trustedClientsLimitOrdersWithTrades: Collection<LimitOrderWithTrades>): Boolean {
```
rocketmannot: suggest to rename, currently looks like we are checking this is trustedClientEvent , but in fact we are checking that data is not empty, the same for `isClientEvent`

src/main/kotlin/com/lykke/matching/engine/order/ExecutionEventSender.kt
```diff
@@ -0,0 +1,96 @@
...
+                           private val messageSender: MessageSender,
+                           private val clientLimitOrdersQueue: BlockingQueue<LimitOrdersReport>,
+                           private val trustedClientsLimitOrdersQueue: BlockingQueue<LimitOrdersReport>,
+                           private val rabbitSwapQueue: BlockingQueue<MarketOrderWithTrades>,
+                           private val lkkTradesQueue: BlockingQueue<List<LkkTrade>>,
+                           private val genericLimitOrderService: GenericLimitOrderService,
+                           private val orderBookQueue: BlockingQueue<OrderBook>,
+                           private val rabbitOrderBookQueue: BlockingQueue<OrderBook>) {
+
+    fun generateSequenceNumbers(executionContext: ExecutionContext): SequenceNumbersWrapper {
```
rocketmannot: suggest to extract this method to separate `SequenceNumberUtils` or create method in  `messageSequenceNumberHolder` that will return `SequenceNumbersWrapper`. Currently it looks like not resposibility of `ExecutionEventSender` to hold logic of sequence number generator

src/main/kotlin/com/lykke/matching/engine/order/ExecutionConfirmationService.kt
```diff
@@ -0,0 +1,22 @@
+package com.lykke.matching.engine.order
+
+import com.lykke.matching.engine.order.transaction.ExecutionContext
+import com.lykke.matching.engine.messages.MessageWrapper
+import org.springframework.stereotype.Component
+
+@Component
+class ExecutionConfirmationService(private val executionPersistenceService: ExecutionPersistenceService,
```
rocketmannot: suggest to rename `ExecutionDataApplyService`, confirmation sounds like we have  to eval predicate and confirm basing on it - to apply data or not

src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
```diff
@@ -122,15 +122,19 @@ class MessageProcessor(config: Config, messageRouter: MessageRouter, application
...
         val genericLimitOrdersCancellerFactory = applicationContext.getBean(GenericLimitOrdersCancellerFactory::class.java)
+        val executionContextFactory = applicationContext.getBean(ExecutionContextFactory::class.java)
 
         this.cashOperationService = applicationContext.getBean(CashOperationService::class.java)
         this.cashInOutOperationService = applicationContext.getBean(CashInOutOperationService::class.java)
         this.reservedCashInOutOperationService = applicationContext.getBean(ReservedCashInOutOperationService::class.java)
         this.cashTransferOperationService = applicationContext.getBean(CashTransferOperationService::class.java)
         this.cashSwapOperationService = applicationContext.getBean(CashSwapOperationService::class.java)
-        this.singleLimitOrderService = SingleLimitOrderService(genericLimitOrderProcessorFactory)
+        this.singleLimitOrderService = SingleLimitOrderService(executionContextFactory,
```
rocketmannot: looks like `SingleLimitOrderService` now can be turned to bean itself

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
```diff
@@ -115,40 +123,68 @@ class WalletOperationsProcessor(private val balancesHolder: BalancesHolder,
...
+                                  assetBalance: AssetBalance) {
+
+    val assetId = assetBalance.asset
+    val clientId = wallet.clientId
+    val originBalance = assetBalance.balance
+    val originReserved = assetBalance.reserved
+    var balance = originBalance
+    var reserved = originReserved
 
     fun apply(): Wallet {
```
rocketmannot: `AssetBalance` is already mutable ? why we can not just set new data directly to AssetBalance ?

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
```diff
@@ -115,40 +123,68 @@ class WalletOperationsProcessor(private val balancesHolder: BalancesHolder,
...
     }
-}
 
-private abstract class AbstractChangedAssetBalance(val assetId: String,
-                                                   val clientId: String,
-                                                   val originBalance: BigDecimal,
-                                                   val originReserved: BigDecimal) {
-    var balance = originBalance
-    var reserved = originReserved
+    private fun isBalanceUpdateNotNeeded(changedAssetBalance: ChangedAssetBalance): Boolean {
```
rocketmannot: we can move this method to `ChangedAssetBalance`

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
```diff
@@ -115,40 +123,68 @@ class WalletOperationsProcessor(private val balancesHolder: BalancesHolder,
...
-        return updates.values.toList()
+        return rabbitClientBalanceUpdatesByClientIdAndAssetId.values.toList()
     }
 
-    private fun defaultChangedAssetBalance(operation: WalletOperation): ChangedAssetBalance {
-        val walletAssetBalance = balancesUpdater.getWalletAssetBalance(operation.clientId, operation.assetId)
-        return ChangedAssetBalance(walletAssetBalance.wallet, walletAssetBalance.assetBalance)
+    override fun getAvailableBalance(clientId: String, assetId: String): BigDecimal {
+        val balance = getChangedCopyOrOriginalAssetBalance(clientId, assetId)
+        return if (balance.reserved > BigDecimal.ZERO)
```
rocketmannot: balance.reserved can be < BigDecimal.ZERO ?

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
```diff
@@ -75,38 +75,46 @@ class WalletOperationsProcessor(private val balancesHolder: BalancesHolder,
...
-            val transactionChangedAssetBalance = it.value
-            clientIds.add(transactionChangedAssetBalance.clientId)
-            val update = updates.getOrPut(key(transactionChangedAssetBalance)) {
-                ClientBalanceUpdate(transactionChangedAssetBalance.clientId,
-                        transactionChangedAssetBalance.assetId,
-                        transactionChangedAssetBalance.changedAssetBalance.originBalance,
-                        transactionChangedAssetBalance.balance,
-                        transactionChangedAssetBalance.changedAssetBalance.originReserved,
-                        transactionChangedAssetBalance.reserved)
+        changedAssetBalances.forEach {
```
rocketmannot: please decompose this method, suggest to divide it on parts `getChangedAssetBalances`, `validateAssetBalances`,  `applyChangedAssetBalancesToCurrentTransactionData`, `generateRabbitMqData`

src/main/kotlin/com/lykke/matching/engine/services/GenericStopLimitOrderService.kt
barsukovskii: renamed

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContextFactory.kt
barsukovskii: Renamed

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionOrderBooksHolder.kt
barsukovskii: Renamed.
We use not only updatedOrder; updatedOrderBook is used too

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionBalancesHolder.kt
barsukovskii: changedWalletsByClientId can contain not changed assetBalances also. This map contains only changed asset balances.

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
barsukovskii: Renamed
To separate responsibilities here is not the best idea as it reduces performance - we need to loop through array twice

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
barsukovskii: answered in the next comment answer

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
barsukovskii: I agree but it is expensive operation to scan orderbook and calculate changed (and removed) orders. That is why this action is implemented outside and performed during execution operation

Or you offer to add additional parameters to this method: ordersToCancel, ordersToReplace, etc. ?

src/main/kotlin/com/lykke/matching/engine/order/process/StopOrderBookProcessor.kt
barsukovskii: This method gets and removes order, that is why it is named 'poll'. And Queue.poll doesn't imply any waiting

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
barsukovskii: yes, fixed

src/main/kotlin/com/lykke/matching/engine/order/process/LimitOrderProcessor.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/order/cancel/GenericLimitOrdersCanceller.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/order/cancel/AbstractLimitOrdersCanceller.kt
barsukovskii: yes, i've planned to refactor it later in the frame of another task. This class and all connected ones will be removed at all

src/main/kotlin/com/lykke/matching/engine/order/cancel/AbstractLimitOrdersCanceller.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/order/ExecutionEventSender.kt
barsukovskii: renamed and moved to utils class

src/main/kotlin/com/lykke/matching/engine/order/ExecutionEventSender.kt
barsukovskii: created ExecutionEventsSequenceNumbersGenerator
moved generateSequenceNumbers to ExecutionEventsSequenceNumbersGenerator
created ExecutionEventUtils
moved isThereTrustedClientEvent & isThereClientEvent to ExecutionEventUtils

src/main/kotlin/com/lykke/matching/engine/order/ExecutionConfirmationService.kt
barsukovskii: renamed

src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
barsukovskii: yes, it will be done in the frame of separate task

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
barsukovskii: we must set changes to wallet - there can be cases when wallet doesn't contain changed asset balance - in this case we lose balance change

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
barsukovskii: moved and renamed to 'isChanged'

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
barsukovskii: Ideally - no, but in theory there can be cases (as a result of some bug or incorrect trusted clients setting changes)

src/main/kotlin/com/lykke/matching/engine/balance/WalletOperationsProcessor.kt
barsukovskii: changed. Added methods processChangedAssetBalance and generateEventData ( I guess it is not good idea to name method 'generateRabbitMqData' - here we don't know where events will be sent, also renamed rabbitClientBalanceUpdatesByClientIdAndAssetId -> clientBalanceUpdatesByClientIdAndAssetId)
processChangedAssetBalance calls generateEventData 
processChangedAssetBalance is called for each element of changedAssetBalances

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
rocketmannot: agree, please add requered ordersToCancel, ordersToReplace, parameters to be sure user have not forgotten to set them

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
rocketmannot: suggest to remove this methods as we will have parameters in `setOrderBook` method

src/main/kotlin/com/lykke/matching/engine/order/transaction/AbstractTransactionOrderBooksHolder.kt
barsukovskii: Reviewed and decided to leave it as is, because:
-method 'setOrderBook' is called by cancellation services for each changed order book and these services have separate lists of cancelled and completed orders of all order books, so they call setOrderBook many times and addCompletedOrders/addCancelledOrders once

-other services don't use the method 'setOrderBook' at all - they get orderBook with the method 'getChangedOrderBookCopy' and change it

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
```diff
@@ -79,10 +68,6 @@ class MarketOrderService @Autowired constructor(
     private var logCount = 100
```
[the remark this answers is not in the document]
barsukovskii: It should be done in a separate task, as it is not connected to current one and not critical (MarketOrderService is used by single thread)

PR #215, opened 2018-11-02
LWDEV-9805 Monitor Rabbit Mq queue size, when size reaches threshold - turn on maintenance mode

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/config/spring/AppConfiguration.kt
```diff
@@ -75,6 +78,24 @@ open class AppConfiguration: SchedulingConfigurer {
                         config = config.me.aliveStatus)
     }
 
+    @Bean
+    open fun inputQueueSizeChecker(@InputQueue namesToInputQueues: Map<String, BlockingQueue<*>>): QueueSizeHealthChecker {
+        return QueueSizeHealthChecker(
```
barsukovskii: Now we can remove InputQueueSizeHealthChecker

rocketmannot: for input queues we are using different values for max queue size and recover queues size - thus we have two queue health checkers in context

src/main/kotlin/com/lykke/matching/engine/config/spring/QueueConfig.kt
```diff
@@ -56,36 +61,43 @@ open class QueueConfig {
     }
 
     @Bean
+    @RabbitQueue
```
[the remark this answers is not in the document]
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/config/spring/AppConfiguration.kt
barsukovskii: I mean the class InputQueueSizeHealthChecker. As I understood now we don't need it

PR #220, opened 2018-11-15
LWDEV-9804 GET /client/{clientId}/orders validations

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/web/controllers/ClientOrdersController.kt
```diff
@@ -22,14 +26,47 @@ class ClientOrdersController {
...
 
-    @GetMapping("client/{clientId}/orders", produces = [MediaType.APPLICATION_JSON_VALUE])
-    @ApiOperation("Endpoint to access orders from current order book for given client")
-    fun getOrders(@PathVariable("clientId") clientId: String,
-                  @RequestParam(required = false) assetPair: String?,
-                  @RequestParam(required = false) isBuy: Boolean?): ClientOrdersDto {
-        val limitOrders = genericLimitOrderService.searchOrders(clientId, assetPair, isBuy)
-        val stopOrders = genericStopLimitOrderService.searchOrders(clientId, assetPair, isBuy)
+    @Autowired
+    private lateinit var assetPairsCache: AssetPairsCache
```
barsukovskii: Let's use AssetsPairsHolder instead of AssetPairsCache

src/main/kotlin/com/lykke/matching/engine/web/controllers/ClientOrdersController.kt
```diff
@@ -22,14 +26,47 @@ class ClientOrdersController {
...
+
+    @Autowired
+    private lateinit var balancesHolder: BalancesHolder
+
+    @GetMapping("wallet/{walletId}/orders", produces = [MediaType.APPLICATION_JSON_VALUE])
```
barsukovskii: path was changed (client -> wallet) but it is not mentioned in the task. Please add info to there

src/main/kotlin/com/lykke/matching/engine/web/controllers/ClientOrdersController.kt
```diff
@@ -22,14 +26,47 @@ class ClientOrdersController {
...
+    }
+
+    @ExceptionHandler
+    private fun handleIllegalArgumentException(request: HttpServletRequest, exception: IllegalArgumentException): ResponseEntity<String> {
+        return ResponseEntity(exception.message ?: "", HttpStatus.BAD_REQUEST)
     }
+
+    @ExceptionHandler
+    private fun handleWalletNotFoundException(request: HttpServletRequest, exception: WalletNotFoundException): ResponseEntity<String> {
+        return ResponseEntity("Wallet is not found - there is no balances data for given wallet", HttpStatus.NOT_FOUND)
```
barsukovskii: Message should be about orders but not balances

PR #229, opened 2018-11-28
LWDEV-10197 Configurable history persistence: added flag to switch on/off history persistence

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/config/spring/MarketHistoryConfig.kt
```diff
@@ -0,0 +1,76 @@
...
+                                             @Value("\${azure.candles.table:#{null}}")
+                                             candlesTable: String?,
+                                             @Value("\${azure.hour.candles.table:#{null}}")
+                                             hourCandlesTable: String?)
+            : LimitOrderDatabaseAccessor? {
+        if (config.me.disableMarketHistory == true) {
+            return null
+        }
+        return AzureLimitOrderDatabaseAccessor(connectionString = config.me.db.hLiquidityConnString,
+                bestPricesTable = bestPricesTable!!, candlesTable = candlesTable!!, hourCandlesTable = hourCandlesTable!!)
```
rocketmannot: suggest to make check for non null for these parameters as well

src/main/kotlin/com/lykke/matching/engine/config/spring/MarketHistoryConfig.kt
```diff
@@ -0,0 +1,76 @@
...
+    @Autowired
+    private lateinit var config: Config
+
+    @Bean
+    open fun azureHistoryTicksDatabaseAccessor(@Value("\${application.tick.frequency:#{null}}")
+                                               frequency: Long?)
+            : HistoryTicksDatabaseAccessor? {
+        if (config.me.disableMarketHistory == true) {
+            return null
+        }
```
rocketmannot: suggest to check in `else` branch that `frequency` is not null - if null - throw `IllegalStateException` with message, in this case we will have meaningful message about misconfiguration instead of NPE.

src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
```diff
@@ -140,45 +130,52 @@ class MessageProcessor(config: Config, messageRouter: MessageRouter, application
...
         this.quotesUpdateHandler = applicationContext.getBean(QuotesUpdateHandler::class.java)
 
         processedMessagesCache = applicationContext.getBean(ProcessedMessagesCache::class.java)
         servicesMap = initServicesMap()
 
-        if (!isLocalProfile) {
-            this.bestPriceBuilder = fixedRateTimer(name = "BestPriceBuilder", initialDelay = 0, period = config.me.bestPricesInterval) {
-                limitOrderDatabaseAccessor.updateBestPrices(genericLimitOrderService.buildMarketProfile())
-            }
+        if (!isLocalProfile && config.me.disableMarketHistory != true) {
```
rocketmannot: suggest to move this logic out of message processor. also suggest to start fixRateTimers from `@PostConstruct` or even better use `@Scheduled` methods inside services

src/main/kotlin/com/lykke/matching/engine/config/spring/MarketHistoryConfig.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/config/spring/MarketHistoryConfig.kt
barsukovskii: fixed

src/main/kotlin/com/lykke/matching/engine/messages/MessageProcessor.kt
barsukovskii: fixed

PR #239, opened 2018-12-24
LWDEV-9019 Expiry time for limit orders (GTD orders)

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
```diff
@@ -0,0 +1,44 @@
...
+            return false
+        }
+        return ordersQueue.add(order)
+    }
+
+    fun removeIfOrderHasExpiryTime(order: LimitOrder): Boolean {
+        if (!order.hasExpiryTime()) {
+            return false
+        }
+        return ordersQueue.remove(order)
```
rocketmannot: for priority queue remove in some cases can iterate over all elements in the quee. Maybe we can add HashSet of limit orders that should be removed from the queue. `getExpiredOrdersExternalIds` perform check `if(setOfRemovedOrders.remove(externalId))` - just skip that order ? If we add new HashSet field - we need to add synchronization to accessor methods

src/main/kotlin/com/lykke/matching/engine/services/GenericLimitOrderService.kt
```diff
@@ -34,6 +36,9 @@ class GenericLimitOrderService @Autowired constructor(private val orderBookDatab
     }
 
     fun update() {
+        limitOrdersMap.values.forEach {
+            expiryOrdersQueue.removeIfOrderHasExpiryTime(it)
```
rocketmannot: why not clear `expiryOrdersQueue` completely instead of removing orders with expiry time one by one ?

src/main/kotlin/com/lykke/matching/engine/order/process/StopLimitOrderProcessor.kt
```diff
@@ -55,7 +55,10 @@ class StopLimitOrderProcessor(private val limitOrderInputValidator: LimitOrderIn
     private fun performBusinessValidation(orderContext: StopLimitOrderContext): OrderValidationResult {
         if (orderContext.limitVolume != null) {
```
rocketmannot: business validation is not performed if `orderContext.limitVolume == null` - so expired order will not be cancelled in this case

src/main/kotlin/com/lykke/matching/engine/order/process/StopLimitOrderProcessor.kt
```diff
@@ -55,7 +55,10 @@ class StopLimitOrderProcessor(private val limitOrderInputValidator: LimitOrderIn
...
             try {
-                stopOrderBusinessValidator.performValidation(calculateAvailableBalance(orderContext), orderContext.limitVolume)
+                stopOrderBusinessValidator.performValidation(calculateAvailableBalance(orderContext),
+                        orderContext.limitVolume,
```
rocketmannot: now we can get limit volume having just order. Suggest to move `calculateLimitVolume` from `StopLimitOrderContext` to util class and calculate limit volume when needed - look like it is cheap to calculate each time

src/main/kotlin/com/lykke/matching/engine/daos/LimitOrder.kt
```diff
@@ -81,4 +87,12 @@ class LimitOrder(id: String,
         origin.lastMatchTime = lastMatchTime
         origin.price = price
     }
+
+    fun hasExpiryTime(): Boolean {
+        return timeInForce == OrderTimeInForce.GTD && expiryTime != null
+    }
+
+    fun isExpired(date: Date): Boolean {
+        return hasExpiryTime() && !expiryTime!!.after(date)
```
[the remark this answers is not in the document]
barsukovskii: I decided that expiryTime=date means that order is exired

src/main/kotlin/com/lykke/matching/engine/incoming/parsers/impl/SingleLimitOrderContextParser.kt
```diff
@@ -136,6 +127,8 @@ class SingleLimitOrderContextParser(val assetsPairsHolder: AssetsPairsHolder,
                 lowerPrice = if (message.hasLowerPrice()) BigDecimal.valueOf(message.lowerPrice) else null,
                 upperLimitPrice = if (message.hasUpperLimitPrice()) BigDecimal.valueOf(message.upperLimitPrice) else null,
                 upperPrice = if (message.hasUpperPrice()) BigDecimal.valueOf(message.upperPrice) else null,
-                previousExternalId = null)
+                previousExternalId = null,
+                timeInForce = if (message.hasTimeInForce()) OrderTimeInForce.getByExternalId(message.timeInForce) else null,
```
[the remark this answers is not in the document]
barsukovskii: Input validation is performed after this limit order object creation. I guess it is wrong and should be refactored in the frame of separate refactoring task - we should validate proto object instead of business LimitOrder object as input data.
Also there is the same situation with other enums here. So, let's leave it now as is.

src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.lykke.matching.engine.daos.LimitOrder
+import org.springframework.stereotype.Component
+import java.util.Date
+import java.util.PriorityQueue
+
+@Component
+class ExpiryOrdersQueue {
+
+    private val ordersQueue = PriorityQueue<LimitOrder> { o1, o2 ->
```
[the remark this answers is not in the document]
barsukovskii: decided to switch to usage of ConcurrentHashMap. Now `getExpiredOrdersExternalIds` just filters this map values each time to return expired orders ids. It is not performed in main business thread, so this overhead is insignificant

src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
```diff
@@ -0,0 +1,44 @@
...
+        var result = o1.expiryTime!!.compareTo(o2.expiryTime!!)
+        if (result == 0) {
```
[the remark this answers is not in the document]
barsukovskii: removed this - switched to usage of ConcurrentHashMap

src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
barsukovskii: removed this - switched to usage of ConcurrentHashMap

src/main/kotlin/com/lykke/matching/engine/services/GenericLimitOrderService.kt
barsukovskii: I decided not to add additional method `clear` for only this method `update` and that it is better to use api method. `update` is only performed during starting ME, so it is not a problem

src/main/kotlin/com/lykke/matching/engine/order/process/StopLimitOrderProcessor.kt
barsukovskii: limitVolume=null means that incoming order is invalid, so we don't have to perform business validation

src/main/kotlin/com/lykke/matching/engine/order/process/StopLimitOrderProcessor.kt
barsukovskii: `limitVolume` should be always calculated before business validation. Business validation is always performed (except when input validation is failed, but this validation should be moved to preprocessor, there is 'todo'). That is why I think it is not a problem to calculate this value at once

src/main/kotlin/com/lykke/matching/engine/daos/LimitOrder.kt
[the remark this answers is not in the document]
rocketmannot: suggest adding small text on the wiki in ME space about different types of orders that may automatically be canceled

src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
barsukovskii: It is not a problem to send the same order several times - it doesn't lead to any error.
I see no reason to make `getExpiredOrdersExternalIds` synchronized. It just returns list of id of current expired orders

src/main/kotlin/com/lykke/matching/engine/order/process/StopLimitOrderProcessor.kt
rocketmannot: Had conversation. `limitVolume` is calculable value - calculated based on required input fields, if input validation passed - this field should not be `null`

src/main/kotlin/com/lykke/matching/engine/order/ExpiryOrdersQueue.kt
rocketmannot: decided to have weekly consistent results for `getExpiredOrdersExternalIds`, suggest adding a comment that this method is weekly consistent. Also decided that expiry orders should be removed from business thread only. - there can be duplicated orders - but currently, in limit order service they are ignored.

PR #244, opened 2019-01-04
LWDEV-10294 Make messageProcessor a spring bean

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/performance/PerformanceStatsHolder.kt
```diff
@@ -8,6 +8,30 @@ class PerformanceStatsHolder {
 
     private var statsMap = HashMap<Byte, PerformanceStats>()
 
+    fun addMessage(type: Byte,
+                           startTimestamp: Long,
```
barsukovskii: format

rocketmannot: fixed

PR #245, opened 2019-01-10
LWDEV-9085 Immediate Or Cancel (IoC) Limit Order

barsukovskii:


src/test/kotlin/com/lykke/matching/engine/services/LimitOrderServiceTest.kt
```diff
@@ -1814,4 +1814,67 @@ class LimitOrderServiceTest : AbstractTest() {
         assertEquals(1, event.orders.size)
         assertEquals(OutgoingOrderStatus.CANCELLED, event.orders.single().status)
     }
+
+    @Test
+    fun testImmediateOrCancelOrderWithTrades() {
```
rocketmannot: Suggest to add case when order has IOC and matched

barsukovskii: Added

PR #246, opened 2019-01-10
LWDEV-10110 Refactor the implementation of limit orders cancel operations

barsukovskii:
Remove obsolete class GenericLimitOrdersCanceller and all connected ones, use ExecutionContext to perform cancellation operations


src/main/kotlin/com/lykke/matching/engine/order/process/common/CancelRequest.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.lykke.matching.engine.daos.LimitOrder
+import com.lykke.matching.engine.deduplication.ProcessedMessage
+import com.lykke.matching.engine.messages.MessageType
+import com.lykke.matching.engine.messages.MessageWrapper
+import org.apache.log4j.Logger
+import java.util.Date
+
+class CancelRequest(val limitOrders: Collection<LimitOrder>,
+                    val stopLimitOrders: Collection<LimitOrder>,
+                    val assetPairId: String?,
```
rocketmannot: is `assetPairId` - needed ?

src/main/kotlin/com/lykke/matching/engine/order/process/common/LimitOrdersCancelExecutorImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+            return executionDataApplyService.persistAndSendEvents(messageWrapper, executionContext)
+        }
+    }
+
+    private fun createAssetPairsByIdMapForOrders(orders: Collection<LimitOrder>): Map<String, AssetPair> {
+        return orders.asSequence()
+                .map { it.assetPairId }
+                .toSet()
+                .mapNotNull { assetsPairsHolder.getAssetPairAllowNulls(it) }
+                .groupBy { it.assetPairId }
```
rocketmannot: As i understand asset pair assetPairId is unique - so we dont need to perform grouping here ?

src/main/kotlin/com/lykke/matching/engine/order/process/common/CancelRequest.kt
barsukovskii: No. Fixed

src/main/kotlin/com/lykke/matching/engine/order/process/common/LimitOrdersCancelExecutorImpl.kt
barsukovskii: Yes, but we group here to return map assetPairId->AssetPair

PR #248, opened 2019-01-11
LWDEV-10222 Stop-limit not converted to LO after stop price reached

barsukovskii:
create LO copy based on StopLO to execute after stop price reached


src/main/kotlin/com/lykke/matching/engine/database/common/PersistenceManagerFactoryImpl.kt
```diff
@@ -42,21 +42,27 @@ class PersistenceManagerFactoryImpl(private val balancesDatabaseAccessorsHolder:
...
-                        redisConnection.get(),
-                        config,
-                        currentTransactionDataHolder,
-                        performanceStatsHolder
-                )
-            }
+            Storage.Redis ->
+                createRedisPersistenceManager(redisConnection.get(), ordersPersistInSecondaryDbStrategy.get())
+            Storage.RedisWithoutOrders ->
+                createRedisPersistenceManager(redisConnection.get(), null)
```
rocketmannot: as i understand we should use version from left side ?

PR #254, opened 2019-02-11
LWDEV-10404 Perform client order spread validation during matching

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/matching/MatchingEngine.kt
```diff
@@ -100,6 +100,11 @@ class MatchingEngine(private val genericLimitOrderService: GenericLimitOrderServ
                     continue
                 }
                 if (order.clientId == limitOrderOrigin.clientId) {
+                    if (order.takePrice() != null) {
```
rocketmannot: As i understand `order.takePrice() != null` - means we have Market order - suggest to add helper method `isMarketOrder()`

barsukovskii: `order.takePrice() != null` is limit order.
fixed: added method `isMarketOrder(order: Order)` to `MatchingEngine`.

PR #255, opened 2019-02-11
LWDEV-10406 One thread for all throttling loggers

barsukovskii:
Currently each ThrottlingLogger instance creates a separate thread to clear their logged errors cache.
Changed kotlin-utils lib - added single thread with one common logged errors cache for all loggers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,7 @@ repositories {
 }
 
 ext {
-    kotlin_utils_version = '0.0.8-SNAPSHOT'
+    kotlin_utils_version = '0.0.9-SNAPSHOT'
     log4j_version = '1.2.17'
     protobuf_version = '3.6.0'
     azure_storage_version = '4.0.0'


build.gradle
```diff
@@ -32,7 +32,7 @@ repositories {
 }
 
 ext {
-    kotlin_utils_version = '0.0.8-SNAPSHOT'
+    kotlin_utils_version = '0.0.9-SNAPSHOT'
```
rocketmannot: why do we use SNAPSHOT - version?

barsukovskii: I can't publish not SNAPSHOT artefact to https://oss.sonatype.org/content/repositories/snapshots/. Perhaps it is connected with rights of my account I created for this library once )

rocketmannot: suggest to use the lates version

barsukovskii: the latest one is not compatible with log4j in this version

PR #259, opened 2019-02-15
LWDEV-10422 Generate RMQ data in a separate thread

rocketmannot:


src/main/kotlin/com/lykke/matching/engine/order/ExecutionEventSender.kt
```diff
@@ -10,8 +10,11 @@ import com.lykke.matching.engine.services.MessageSender
...
 import com.lykke.matching.engine.utils.event.isThereClientEvent
 import com.lykke.matching.engine.utils.event.isThereTrustedClientEvent
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.core.task.TaskExecutor
 import org.springframework.stereotype.Component
 import java.util.concurrent.BlockingQueue
+import javax.annotation.PostConstruct
 
 @Component
 class ExecutionEventSender(private val messageSender: MessageSender,
```
barsukovskii: This sender is used only for orders (limit-, multi-, market- and cancellation) and not used for cash operations. I propose to implement generation RMQ data in a separate thread for them too at once. So, there should be common queue for all messages to generate RMQ data (without using current `RabbitEventData` as it is specific for execution event)

rocketmannot: fixed - added async handling for rest of event types

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionOrderBooksHolder.kt
```diff
@@ -97,12 +96,20 @@ class CurrentTransactionOrderBooksHolder(private val genericLimitOrderService: G
         }
     }
 
-    private fun processChangedOrderBookSide(orderBookCopy: AssetOrderBook, isBuySide: Boolean, date: Date) {
-        val assetPairId = orderBookCopy.assetPairId
-        val price = if (isBuySide) orderBookCopy.getBidPrice() else orderBookCopy.getAskPrice()
+    private fun processChangedOrderBookSide(orderBook: AssetOrderBook,
```
barsukovskii: Is there any profit of passing array instead of queue ? If it is then I propose not to create copy of order book before invocation of processChangedOrderBookSide

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/CashTransferEventSenderService.kt
```diff
@@ -0,0 +1,42 @@
...
+
+@Service
+class CashTransferEventSenderService(
+        private val cashTransferEventData: BlockingQueue<CashTransferEventData>,
+        private val eventSenders: List<SpecializedCashTransferEventSender>,
+        @Qualifier("rabbitPublishersThreadPool")
+        private val rabbitPublishersThreadPool: TaskExecutor) {
+
+    @PostConstruct
+    private fun init() {
```
barsukovskii: I think it is better to add abstract class for sender services and move this method to parent to deduplicate (perhaps also some other ones, as CashInOutEventSenderService and CashTransferEventSenderService look similar)
I would create common interface (with generic type) for all event senders instead of SpecializedCashInOutEventSender, SpecializedCashTransferEventSender and SpecializedExecutionEventSender and common abstract implementation for new format event senders

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/ExecutionEventSenderImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+@Component
+class ExecutionEventSenderImpl(private val messageSender: MessageSender) : SpecializedExecutionEventSender {
+
+    override fun sendEvent(executionData: ExecutionData) {
+        val executionContext = executionData.executionContext
+        sendTrustedClientsExecutionEventIfNeeded(executionContext, executionData)
+        sendClientsExecutionEventIfNeeded(executionContext, executionData)
+    }
+
+    fun sendClientsExecutionEventIfNeeded(executionContext: ExecutionContext, executionEventData: ExecutionData) {
```
barsukovskii: propose to make `sendClientsExecutionEventIfNeeded` and `sendTrustedClientsExecutionEventIfNeeded` a private and with single parameter - executionEventData: ExecutionData.

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/OldFormatExecutionEventSender.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    fun sendBalanceUpdateEvent(executionContext: ExecutionContext) {
+        executionContext.walletOperationsProcessor.sendNotification(executionContext.requestId, executionContext.messageType.name, executionContext.messageId)
+    }
+
+    fun sendTrustedClientsExecutionEventIfNeeded(executionContext: ExecutionContext) {
+        val trustedClientsLimitOrdersWithTrades = executionContext.getTrustedClientsLimitOrdersWithTrades().toMutableList()
+
+        if (isThereTrustedClientEvent(trustedClientsLimitOrdersWithTrades)) {
+            trustedClientsLimitOrdersQueue.put(LimitOrdersReport(executionContext.messageId, trustedClientsLimitOrdersWithTrades.toMutableList()))
```
barsukovskii: redundant `toMutableList` invocation

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
```diff
@@ -118,6 +105,16 @@ class MarketOrderService @Autowired constructor(
                 now,
                 LOGGER)
 
+        try {
+            marketOrderValidator.performValidation(order, getOrderBook(order), feeInstruction, feeInstructions)
```
barsukovskii: Perform validation before creating context, otherwise it leads to NPE in case of unknown asset pair (please add unit test for this case)
Also I think it is not correct to create context even for invalid orders that will not be executed exactly (as I understood, currently it is created only for sending rejected order event and i'm not sure it is correct)

src/test/kotlin/com/lykke/matching/engine/config/TestApplicationContext.kt
```diff
@@ -546,12 +555,13 @@ open class TestApplicationContext {
...
-                                          cashTransferOperationBusinessValidator: CashTransferOperationBusinessValidator, messageSequenceNumberHolder: MessageSequenceNumberHolder,
-                                          messageSender: MessageSender): CashTransferOperationService {
-        return CashTransferOperationService(balancesHolder, notification, dbTransferOperationQueue, feeProcessor,
-                cashTransferOperationBusinessValidator, messageSequenceNumberHolder, messageSender)
+                                          cashTransferOperationBusinessValidator: CashTransferOperationBusinessValidator,
+                                          messageSequenceNumberHolder: MessageSequenceNumberHolder,
+                                          cashTransferOperationEventSender: CashTransferEventSender): CashTransferOperationService {
+        return CashTransferOperationService(balancesHolder,  dbTransferOperationQueue, feeProcessor,
+                cashTransferOperationBusinessValidator, messageSequenceNumberHolder, cashTransferOperationEventSender)
```
barsukovskii: wrong type of the last parameter

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/ExecutionEventSenderImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+                    executionContext.date,
+                    executionContext.messageType,
+                    executionContext.walletOperationsProcessor.getClientBalanceUpdates(),
+                    clientsLimitOrdersWithTrades,
+                    executionContext.marketOrderWithTrades))
+        }
+    }
+
+    fun sendTrustedClientsExecutionEventIfNeeded(executionContext: ExecutionContext, executionEventData: ExecutionData) {
+        val trustedClientsLimitOrdersWithTrades = executionContext.getTrustedClientsLimitOrdersWithTrades().toMutableList()
```
barsukovskii: `toList` can be called instead of `toMutableList`

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/OldFormatExecutionEventSender.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    fun sendClientsExecutionEventIfNeeded(executionContext: ExecutionContext) {
+        val clientsLimitOrdersWithTrades = executionContext.getClientsLimitOrdersWithTrades().toList()
```
barsukovskii: I propose to invoke `toMutableList `here and not to invoke `toMutableList `at line 43 to avoid recreation of the list

src/main/kotlin/com/lykke/matching/engine/order/transaction/CurrentTransactionOrderBooksHolder.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/CashTransferEventSenderService.kt
rocketmannot: fixed now we have only one SpecializedEvenetSender interface

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/ExecutionEventSenderImpl.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/OldFormatExecutionEventSender.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/services/MarketOrderService.kt
rocketmannot: returned prev impl. after release of Multilimit preprocessors we can be sure that asset pair exists. I think our goal should be to incapsulate outgoing event queues in senders - we will do that after merge of multilimit order context

src/test/kotlin/com/lykke/matching/engine/config/TestApplicationContext.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/ExecutionEventSenderImpl.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/OldFormatExecutionEventSender.kt
rocketmannot: fixed

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/SpecializedEventSender.kt
```diff
@@ -0,0 +1,6 @@
+package com.lykke.matching.engine.outgoing.senders
+
+interface SpecializedEventSender<T> {
+    fun getEventClass(): Class<T>
+    fun sendEvent(event: Any)
```
barsukovskii: It is better to use Generic instead of Any

src/main/kotlin/com/lykke/matching/engine/outgoing/messages/OutgoingEventDataWrapper.kt
```diff
@@ -0,0 +1,3 @@
+package com.lykke.matching.engine.outgoing.messages
+
+class OutgoingEventDataWrapper<T>(val eventClass: Class<T>, val eventData: T)
```
barsukovskii: `eventClass` looks like redundant field - you can use eventData::class.java instead

rocketmannot: removed OutgoingEventDataWrapper

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/SpecializedEventSender.kt
rocketmannot: added `OutgoingEventData` marker interface - usage of generic here is not convenient as in the code we add SpecializedEventSenders to aggregated  one map.

src/test/kotlin/com/lykke/matching/engine/matching/MatchingEngineFeeTest.kt
```diff
@@ -71,6 +76,20 @@ class MatchingEngineFeeTest : MatchingEngineTest() {
         )
     }
 
+    @Test
+    fun test() {
```
barsukovskii: It seems this is a temporary performance test

rocketmannot: removed

PR #264, opened 2019-04-15
LWDEV-10133 Add outgoing event RESERVED_BALANCE_UPDATE

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/specialized/ReservedCashInOutOldEventSender.kt
```diff
@@ -0,0 +1,30 @@
...
+        return ReservedCashInOutEventData::class.java
+    }
+
+    override fun sendEvent(event: ReservedCashInOutEventData) {
+        event.walletOperationsProcessor.sendNotification(event.requestId, MessageType.RESERVED_CASH_IN_OUT_OPERATION.name, event.messageId)
+
+        reservedCashOperationQueue.put(ReservedCashOperation(event.requestId,
+                event.walletOperation.clientId,
+                event.date,
+                NumberUtils.setScaleRoundHalfUp(event.walletOperation.reservedAmount, event.accuracy).toPlainString(),
```
rocketmannot: suggest to move `bigDecimalToString` to NumberUtils and use it here

src/main/kotlin/com/lykke/matching/engine/outgoing/messages/v2/builders/ReservedBalanceUpdateEventBuilder.kt
```diff
@@ -0,0 +1,25 @@
...
+    private var balanceUpdates: List<BalanceUpdate>? = null
+    private var reservedBalanceUpdate: ReservedBalanceUpdate? = null
+
+    override fun getMessageType() = MessageType.RESERVED_BALANCE_UPDATE
+
+    override fun setEventData(eventData: ReservedBalanceUpdateData): EventBuilder<ReservedBalanceUpdateData, ReservedBalanceUpdateEvent> {
+        balanceUpdates = convertBalanceUpdates(eventData.clientBalanceUpdates)
+        reservedBalanceUpdate = ReservedBalanceUpdate(eventData.reservedBalanceUpdateOperation.clientId,
+                eventData.reservedBalanceUpdateOperation.assetId,
+                bigDecimalToString(eventData.reservedBalanceUpdateOperation.reservedAmount)!!)
```
rocketmannot: why `reservedAmount` is not rounder? in old publisher it is rounded

src/main/kotlin/com/lykke/matching/engine/outgoing/senders/impl/specialized/ReservedCashInOutOldEventSender.kt
barsukovskii: current conversion and `bigDecimalToString` can return different values, so I decided to leave it as is for old format backward compatibility - it has been taken from the previous implementation

src/main/kotlin/com/lykke/matching/engine/outgoing/messages/v2/builders/ReservedBalanceUpdateEventBuilder.kt
barsukovskii: the idea of the new contract is to return the same value we got

PR #266, opened 2019-04-15
LWDEV-10476 Pre-trade control: configurable max value and calculated max volume

barsukovskii:


rocketmannot:
Not sure on limit orders - what if 
1) Input limit order (A) passes validation - set to order book
2) Mid Price is changing
3) Limit order(A) is in the order book and has invalid - too big volume according to the current mid price

PR #267, opened 2019-04-17
LWDEV-10509 Order book max size control

barsukovskii:


src/main/kotlin/com/lykke/matching/engine/utils/config/MatchingEngineConfig.kt
```diff
@@ -43,5 +43,6 @@ data class MatchingEngineConfig(
         val disableBlobHistory: Boolean?,
         val disableBestPriceHistory: Boolean?,
         val disableCandlesHistory: Boolean?,
-        val disableHourCandlesHistory: Boolean?
+        val disableHourCandlesHistory: Boolean?,
+        val orderBookMaxTotalSize: Int?
```
rocketmannot: may be we should create a separate section for controls params ?

src/main/kotlin/com/lykke/matching/engine/holders/OrderBookMaxTotalSizeHolder.kt
```diff
@@ -0,0 +1,5 @@
+package com.lykke.matching.engine.holders
+
+interface OrderBookMaxTotalSizeHolder {
```
rocketmannot: why do we need this class?

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContext.kt
```diff
@@ -68,6 +68,10 @@ class ExecutionContext(val messageId: String,
     fun getClientsLimitOrdersWithTrades() = clientLimitOrdersWithTradesByInternalId.values
     fun getTrustedClientsLimitOrdersWithTrades() = trustedClientLimitOrdersWithTradesByInternalId.values
 
+    fun getOrderBookTotalSize(): Int {
+        return orderBooksHolder.getOrderBookTotalSize() + stopOrderBooksHolder.getOrderBookTotalSize()
```
rocketmannot: why we are checking total order book size? looks like currently big order book for one asset can cause trade stop. Maybe we should check the order book size per asset?

src/main/kotlin/com/lykke/matching/engine/services/validators/business/impl/LimitOrderBusinessValidatorImpl.kt
```diff
@@ -11,12 +12,16 @@ import java.math.BigDecimal
...
+    : LimitOrderBusinessValidator {
+
     override fun performValidation(isTrustedClient: Boolean, order: LimitOrder,
                                    availableBalance: BigDecimal,
                                    limitVolume: BigDecimal,
                                    orderBook: AssetOrderBook,
-                                   date: Date) {
+                                   date: Date,
+                                   currentOrderBookTotalSize: Int) {
+        OrderValidationUtils.validateOrderBookTotalSize(currentOrderBookTotalSize, orderBookMaxTotalSizeHolder.get())
```
rocketmannot: I think it is not logically right place for this check, as in fact, input order may be valid. Also, I think we should write a message to the metrics log as soon as we detect this issue. I actually not  sure, how we should resolve it

src/main/kotlin/com/lykke/matching/engine/utils/config/MatchingEngineConfig.kt
rocketmannot: I think on ME startup we should check if our order book size already reached the limit in case it reached - throw an exception. Thus we will prevent the case of misconfiguration - when `orderBookMaxTotalSize` was set too low.

barsukovskii: - moved new parameter to the section 'queueConfig'
- added initial order book total size check

src/main/kotlin/com/lykke/matching/engine/holders/OrderBookMaxTotalSizeHolder.kt
barsukovskii: it will be easier to refactor, for example, if we decide to add an ability to configure this value dynamically in the future

src/main/kotlin/com/lykke/matching/engine/order/transaction/ExecutionContext.kt
barsukovskii: I decided that we don't have to check order book size per asset. The only goal of this check is to prevent memory error. Parameter orderBookMaxTotalSize must be a huge number and in reality this check is highly unlikely be ever failed.

src/main/kotlin/com/lykke/matching/engine/services/validators/business/impl/LimitOrderBusinessValidatorImpl.kt
barsukovskii: I agree, I thought about it too - we don't check any order attribute here.. But I decided to place this check here to keep all validations in one place

Added writing message to the metrics log inside the check method OrderValidationUtils.validateOrderBookTotalSize

